PR #19, opened 2016-08-03
consideration for node

LoganWright:
Seeing how this feels over `[String: String]`

Benefits:

```Swift
let id = try request.params.extract("id") as Int // any node convertible
// or older style
guard let id = request.parameters["id"].int else { throw or something }
```


LoganWright:
After internal discussion, we decided that parameters doesn't make as much sense as part of engine and will be moved to either routing or Vapor.

PR #41, opened 2016-08-26
Shim for url encoding to allow Turnstile to work

edjiang:
I'll open up an issue with more details.

### DIFF
--- a/Sources/HTTP/Models/Request/Request.swift
+++ b/Sources/HTTP/Models/Request/Request.swift
@@ -38,12 +38,12 @@ public final class Request: Message {
         // status-line = HTTP-version SP status-code SP reason-phrase CRL
         var path = uri.path
         if let q = uri.query, !q.isEmpty {
-            let encoded = try percentEncoded(q.bytes).string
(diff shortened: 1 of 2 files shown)


Sources/HTTP/Models/Request/Request.swift
```diff
@@ -38,12 +38,12 @@ public final class Request: Message {
...
-            let encoded = try percentEncoded(q.bytes).string
-            path += "?\(encoded)"
+            let encoded = q.addingPercentEncoding(withAllowedCharacters: .urlQueryAllowed)
+            path += "?\(encoded ?? "")"
         }
         if let f = uri.fragment, !f.isEmpty {
-            let encoded = try percentEncoded(f.bytes).string
-            path += "#\(encoded)"
+            let encoded = f.addingPercentEncoding(withAllowedCharacters: .urlFragmentAllowed)
+            path += "#\(encoded ?? "")"
```
LoganWright: if encoded is `nil`, should we still have the `#`. Same question w/ query and `?`

PR #48, opened 2016-08-31
tls default configs

LoganWright:


LoganWright:
DO NOT MERGE

Pending TLS Certificates as default as opposed to skipping verify.

PR #61, opened 2016-10-08
Changed Response Header comparison to lowercase as common ground

Narsail:
Some Websockets respond with a lowercase “upgrade” or an uppercase
“WebSocket” string. Therefore a broader check for the headers is
necessary.


LoganWright:
Looks good to me, let's get this merged.

PR #71, opened 2016-11-05
Enable nonblocking operation

andreasley:
This PR is part of a series to enable nonblocking operation in Vapor and solve https://github.com/vapor/engine/issues/66

This PR depends on https://github.com/vapor/socks/pull/86.

~~If this PR is accepted, Vapor needs to be updated with https://github.com/vapor/vapor/pull/690 to work again.~~

### DIFF
--- a/Sources/HTTP/Server/ServerError.swift
+++ b/Sources/HTTP/Server/ServerError.swift
@@ -6,5 +6,4 @@ public enum ServerError: Swift.Error {
     case respond(Swift.Error)
     case dispatch(Swift.Error)
     case unknown(Swift.Error)
-    case notSupported
 }
--- a/Sources/HTTP/Server/ServerError.swift
+++ b/Sources/HTTP/Server/ServerError.swift
@@ -6,4 +6,5 @@ public enum ServerError: Swift.Error {
     case respond(Swift.Error)
     case dispatch(Swift.Error)
     case unknown(Swift.Error)
+    case notSupported
 }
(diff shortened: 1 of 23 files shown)


LoganWright:
@andreasley is there an appropriate set of defaults for watchable that we could implement to properly deprecate w/o breaking stream conformance api?

andreasley:
@LoganWright Absolutely; we can just add the following extension:

```swift
extension Stream {
    public func startWatching(on queue: DispatchQueue, handler: @escaping () -> ()) throws {
        throw StreamError.unsupported
    }
    
    public func stopWatching() throws {
        throw StreamError.unsupported
    }
}
```

This way, the placeholder implementations in other `Stream` classes become redundant and can be removed. 

I have added a commit with these changes. Should have done it like this from the start; not sure what I was thinking. :)

Sources/HTTP/Server/ServerError.swift
```diff
@@ -6,4 +6,5 @@ public enum ServerError: Swift.Error {
     case respond(Swift.Error)
     case dispatch(Swift.Error)
     case unknown(Swift.Error)
+    case notSupported
```
LoganWright: We'll have to pull this out unfortunately. We can throw a `respond` error and have a `notSupported` error somewhere else for API preservation.

andreasley: Fine by me, but `.respond` still requires a Swift.Error as parameter. Are there any generic ones available?

Issue #76, opened 2016-11-14
WebSocket connection header case sensitivity


Obbut:
The [check for the 'Upgrade' value in the connection header](https://github.com/vapor/engine/blob/master/Sources/WebSockets/Request%2BWebSockets.swift#L18) is case sensitive. I don't know if this is according to the HTTP spec, but various nginx example configurations send the 'upgrade' value in lowercase.

I think this check should be made case insensitive.

LoganWright:
@Obbut good call!

kandelvijaya:
Lets close this. This one is already resolved.

PR #77, opened 2016-11-14
WebSocket connection header case sensitivity #76

antosj:
Not sure about adding a test for it.


andreasley:
Looks good.
Additionally, according to [RFC6455](https://tools.ietf.org/html/rfc6455#section-4.2.1), the `upgrade` header field value should also be treated as case-insensitive (in `guard headers.upgrade == "websocket"`).

antosj:
Nice catch, @andreasley! Thanks!

andreasley:
Hm, I've just noticed that there's similar code in [Headers+WebSockets.swift](https://github.com/vapor/engine/blob/1f95094ee470408309e98dd56b2251210d6a2a3d/Sources/WebSockets/Connection/Headers%2BWebSockets.swift), but this code, `isWebSocketRequest`, is never called as far as I can tell. 

@LoganWright Think we should fix this one as well or deprecate it?

@antosj Thanks for the fix. Also, I think when checking the `connection` header, comparing the strings would be more correct than the currently used range check, which returns `true` for the value `blaupgradebla`.

andreasley:
@antosj Because we don't actually need to know if the case is identical, we can improve performance and readability by comparing like this: 

```swift
headers.upgrade?.lowercased() == "websocket"
```

antosj:
May be interesting to know that according to [this document](https://github.com/apple/swift/blob/7e68e02b4eaa1cf44037a383129cbef60ea55d67/docs/StringDesign.rst):

> Whether a direct interface is needed at all in Swift, and if so, its form, are TBD. However, we should consider following the lead of Python 3, wherein case conversion also normalizes letterforms. Then one can combine String.toLower() with default comparison to get a case-insensitive comparison:

They might have been implemented this search option exactly like this. But we don't need to lowercase the constant and NSComparisonResult.

LoganWright:
@antosj looks ok, but we do need the range of on Upgrade, I believe more can be included in that key and we just need to see if Upgrade is one of them

andreasley:
@LoganWright My bad, I asked @antosj to replace the range test. Why is it required? I didn't see anything in [RFC 6455](https://tools.ietf.org/html/rfc6455) to that effect.

LoganWright:
@andreasley I don't remember specifically, but I remember adding it. I believe occasionally we were receiving it in unexpected ways, and at this point, I don't want to change it just in case there's services out there where it's important to have contains logic.

I went ahead and added it, merging now :)

PR #78, opened 2016-11-16
Improved cleanup/closing of sockets

andreasley:
With the introduction of nonblocking sockets, some new cases need to be handled. Servers can now potentially be stopped and started again, so they have to properly close their sockets.


Sources/HTTP/Models/Server/HTTP+Server.swift
```diff
@@ -133,14 +140,15 @@ public final class Server<
...
                 try response.onComplete?(stream)
                 if !request.keepAlive {
+                    self?.streams.remove(stream)
                     try stream.close()
                 }
             } catch ParserError.streamEmpty {
+                self?.streams.remove(stream)
                 try? stream.close()
             } catch {
                 errors(.respond(error))
```
LoganWright: we don't need to remove the stream here?

Tests/HTTPTests/HTTPBodyTests.swift
```diff
@@ -2,12 +2,14 @@ import XCTest
...
 import Transport
 import libc
+import SocksCore
 
 class HTTPBodyTests: XCTestCase {
     static var allTests = [
         ("testBufferParse", testBufferParse),
         ("testChunkedParse", testChunkedParse),
         ("testClientStreamUsage", testClientStreamUsage),
+//        ("testReleasingServer", testReleasingServer),
```
LoganWright: Is this test no longer valid?

Sources/HTTP/Models/Server/HTTP+Server.swift
andreasley: I don't think so. The responder or serializer can fail, but that doesn't mean the HTTP connection needs to be closed. This behavior is identical to the blocking version.

Tests/HTTPTests/HTTPBodyTests.swift
andreasley: This test is new. It starts a server, connects to it, stops it and checks if the previously opened connection was correctly closed. For some reason, this test fails about 20% of the time in my environment. I'm not sure why and will have to look at SocksCore again. I think the PR can be merged without the new test.

PR #81, opened 2016-12-07
Re-added error logging if a `StreamError` occurs

andreasley:
Error logging for `StreamError` is still required for debugging purposes (@LoganWright I wrongly omitted that in the previous PR).


LoganWright:
@andreasley good catch, thanks

PR #85, opened 2016-12-17
lastPathComponent made public

eolcott:
It'd be nice to have access to this convenience


LoganWright:
Extremely late to this merge @eolcott but thank you for the contribution!

PR #89, opened 2016-12-27
Ensure that the sender & receiver of a close with an error get the error code and message.

cwoloszynski:
At the sender, sets the Payload Length properly (to get the error code and message delivered).  

At the receiver, extract the error message from the right variable

### DIFF
--- a/Sources/WebSockets/Model/WebSocket.swift
+++ b/Sources/WebSockets/Model/WebSocket.swift
@@ -115,11 +115,11 @@ extension WebSocket {
             do {
                 let frame = try parser.acceptFrame()
                 try received(frame)
-            } catch {
+            } catch { // IGNORE THIS error for now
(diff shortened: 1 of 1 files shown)


Sources/WebSockets/Model/WebSocket.swift
```diff
@@ -237,14 +237,22 @@ extension WebSocket {
         guard state == .open else { return }
         state = .closing
 
+        var payload: Bytes = []
+        if let status = statusCode {
+            payload += status.bytes()
+        }
+        if let reason = reason {
+            payload += reason.toBytes()
```
LoganWright: I think just `reason.bytes` here from `Core`

cwoloszynski:
This needs the fixes in 1.3.5 to pass Travis, so I am going to close this pull and re-create this based on the 1.3.5 release.

PR #90, opened 2016-12-27
Add support for client-specified headers

cwoloszynski:
These changes allow the WebSocket client to include client-specified headers.  This allows using VaporJWT for authorization of web socket clients

### DIFF
--- a/Sources/WebSockets/Client/WebSocket+Client.swift
+++ b/Sources/WebSockets/Client/WebSocket+Client.swift
@@ -4,30 +4,34 @@ import Transport
 import URI
 import HTTP
 
+public enum ResponseError: Swift.Error {
+    case status(Status)
+}
+
 extension WebSocket {
-    public static func background(to uri: String, using client: ClientProtocol.Type = Client<TCPClientStream, Serializer<Request>, Parser<Response>>.self, protocols: [String]? = nil, onConnect: @escaping (WebSocket) throws -> Void) throws {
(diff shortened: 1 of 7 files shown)


Sources/WebSockets/Client/WebSocket+Client.swift
```diff
@@ -50,6 +54,7 @@ extension WebSocket {
         )
         let response = try client.respond(to: request)
 
+        guard response.status == .switchingProtocols else { throw ResponseError.status(response.status) }
```
LoganWright: this is already accounted for in L61, please rm

Sources/WebSockets/Client/WebSocket+Client.swift
```diff
@@ -4,30 +4,34 @@ import Transport
 import URI
 import HTTP
 
+public enum ResponseError: Swift.Error {
```
LoganWright: rm, not necessary

Sources/WebSockets/Client/WebSocket+Client.swift
cwoloszynski: In my case, I am using a Middleware test that is responding with 'unauthorized'.  I am not sure that 'invalidOrUnsupportedStatus' as an error code is as expressive as wrapping the actual status code and providing in the Throw.    I also don't think that it falls under a 'FormatError' classification.

I put it in front of the other tests, since those tests likely only apply if the response is 'switchingProtocols'.  

Can I ask you to reconsider?

Can I suggest replacing the functionality of L61 with my code?

PR #93, opened 2017-01-25
Created a URI extension to make modifying the query keys/values much easier

Ponyboy47:
This allows middleware or really anything with access to a URI object to modify that URI's query with ease. For example, operations can be performed on a uri to add a "cameFrom" query to the uri, like in the LoginRedirectMiddleware ;)

https://github.com/vapor/vapor/pull/803


Sources/URI/Model/URI+Query.swift
```diff
@@ -0,0 +1,149 @@
...
+        if let value = value?.addingPercentEncoding(withAllowedCharacters: .urlQueryAllowed) {
+            newQuery += "=\(value)"
+        }
+        // Get the existing query
+        guard var q = query else {
+            query = newQuery
+            return
+        }
+        // If there are already keys/values in the query, append an ampersand and then the new query
+        if (q.characters.count > 0) {
```
kandelvijaya: I think its better to use `!q.isEmpty` here.

LoganWright:
We have a lot of these conveniences on request, is there a use case for URI in particular outside of that context?

Ponyboy47:
I made the conveniences on URI itself since I was dealing with URI objects for the LoginRedirectMiddleware. I also just think it's better to have the conveniences available at the lowest level possible.

But since the LoginRedirectMiddleware changes got rejected with all the major changes to Vapor 2, I don't think that this merge is necessary anymore.

PR #94, opened 2017-01-27
Fixes URI parsing when bytes contains authority and existingHost is non nil.

cuva:
An issue surfaced when parsing the URI of a proxied HTTP request.

### Regular Query

```
> nc -l 4242

[...]

> curl http://localhost:4242/api/v2/

GET /api/v2/ HTTP/1.1
Host: localhost:4242
User-Agent: curl/7.51.0
Accept: */*

```

### Proxied Query

```
> nc -l 4242

[...]

> http_proxy=http://localhost:4242 curl http://pokeapi.co/api/v2/

GET http://pokeapi.co/api/v2/ HTTP/1.1
Host: pokeapi.co
User-Agent: curl/7.51.0
Accept: */*
Proxy-Connection: Keep-Alive
```

The presence of an `existingHost` isn't enough to determine if the buffer is free of an authority (as seen in the proxied query example above). This change prevents `parseAuthority` from bailing out if an `existingHost` is present by prioritizing the check on the buffer first.

In this instance, a decision has to be made whether the returned authority should come from `existingHost` or from the buffer. Currently the authority from the buffer will be returned. The added test case also asserts that the returned authority originates from the bytes array rather than the optional `existingHost` property.

### DIFF
--- a/Sources/URI/Parser/Parser+Components.swift
+++ b/Sources/URI/Parser/Parser+Components.swift
@@ -39,8 +39,7 @@ extension URIParser {
         authority   = [ userinfo "@" ] host [ ":" port ]
     */
     internal func parseAuthority() throws -> [Byte]? {
-        if let existingHost = existingHost { return existingHost.array }
-        guard try checkLeadingBuffer(matches: .forwardSlash, .forwardSlash) else { return nil }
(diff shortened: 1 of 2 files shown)


LoganWright:
@cuva this looks good to me, thanks for adding this!

PR #136, opened 2017-05-26
properly parse port from host header when applicable

LoganWright:
vapor/vapor#1031


valeriomazzeo:
Not sure it could happen but what if the host is: `username:password@localhost:8080`?

Shall we consider this case to?

LoganWright:
@valeriomazzeo that wouldn't be a host, since the `username:password` is another section of the uri. If that's in the headers, then that would be a bug in the client.

Also, that format has been deprecated in HTTP spec either way, so I don't think we should be adding special cases to support that one 👍

PR #165, opened 2017-08-22
fix dispatch crashing under heavy load

Joannis:


Joannis:
Relies on Vapor/Core https://github.com/vapor/core/pull/65

PR #299, opened 2018-07-13
Fix for HTTPDecoder NIO issue

mura-admin:
In the current design of HTTPDecoder it needs to be initialised within the channelInitializer context as this may be called more than once. Not doing so can result in a crash. To workaround I've submitted a pull request that uses Strings to keep track of the handlers and remove them; instead of using handler references.
 
For reference: https://github.com/apple/swift-nio/issues/506
Example SwiftPM project that will error: https://github.com/carumusan/swift-nio-exc-error


Cellane:
I think we just hit this issue in our current project. Is there any way to verify that this fix is the way to go forward, and hopefully merge & release a fixed version soon?

penny-coin:
Hey @mura-admin, you just merged a pull request, have a coin! 

You now have 2 coins.

PR #306, opened 2018-08-24
Send close header if the request 'keep alive' is false

enricenrich:
Following the `NIOHTTP1Server` example of SwiftNIO, we send the close header if the request "keep alive" is false.

Tested it in the conditions that it was throwing an error (attached below), with this change it's working as expected.

```
[ ERROR ] data received after completed connection: close message ([...]/Sources/Vapor/Logging/Logger+LogError.swift:32)
```

Fixes: https://github.com/vapor/http/issues/305

### DIFF
--- a/Sources/HTTP/Responder/HTTPServer.swift
+++ b/Sources/HTTP/Responder/HTTPServer.swift
@@ -236,8 +236,24 @@ private final class HTTPServerHandler<R>: ChannelInboundHandler where R: HTTPSer
         // a valid request
         var reshead = res.head
         reshead.headers.add(name: "date", value: RFC1123DateCache.shared.currentTimestamp())
+
         if let server = serverHeader {
             reshead.headers.add(name: "server", value: server)
+        }
+
+        // add 'Connection' header if needed
+        let connectionHeaders = reshead.headers[canonicalForm: "connection"].map { $0.lowercased() }
+
+        if !connectionHeaders.contains("keep-alive") && !connectionHeaders.contains("close") {
+            // the user hasn't pre-set either 'keep-alive' or 'close', so we might need to add headers
+            switch reqhead.isKeepAlive {
+            case true:
+                // the request has 'Connection: keep-alive', we should mirror that
+                reshead.headers.add(name: "Connection", value: "keep-alive")
+            case false:
+                // the request has 'Connection: close', we should mirror that
+                reshead.headers.add(name: "Connection", value: "close")
+            }
         }
 
         // begin serializing
--- a/Sources/HTTP/Responder/HTTPServer.swift
+++ b/Sources/HTTP/Responder/HTTPServer.swift
@@ -224,6 +224,11 @@ private final class HTTPServerHandler<R>: ChannelInboundHandler where R: HTTPSer
             self.errorHandler(error)
             ctx.close(promise: nil)
         }
+        res.whenComplete {
+            if !head.isKeepAlive {
+                ctx.close(promise: nil)
+            }
+        }
     }
 
     /// Serializes the `HTTPResponse`.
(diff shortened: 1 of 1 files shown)


enricenrich:
@tanner0101 Based on my test, it was enough just calling `ctx.close(promise: nil)`. It's also how SwiftNIO does this in the example.

About putting this at the end of `whenSuccess`, I just tried it and it also works. 👍

joscdk:
LGTM

enricenrich:
@tanner0101 Makes sense to move the `ctx.close(promise: nil)` to the `whenSuccess` block at the end I think, specially because as you said the `whenFailure` block also calls it.

Do you want me to do the change?

weissi:
Indeed @tanner0101 on top of the actual close (as in the PR) you should also set `connection: close`. NIO’s example server has an example for that: https://github.com/apple/swift-nio/blob/master/Sources/NIOHTTP1Server/main.swift#L36

But tbh that’s probably overly complicated, think setting `connection: close` iff `request.isKeepAlive == false` sounds good enough. The reason the code is more complicated is that it predates `request.isKeepAlive`.

enricenrich:
@tanner0101 @weissi @joscdk I've just pushed two commits to implement the suggestions mentioned on the PR.

The `ctx.close(promise: nil)` is now called at the end of the `whenSuccess` block instead of the `whenComplete` block to avoid being called twice.

And on serializing the `HTTPResponse` we add the `Connection` header if needed.

Sources/HTTP/Responder/HTTPServer.swift
```diff
@@ -236,10 +236,26 @@ private final class HTTPServerHandler<R>: ChannelInboundHandler where R: HTTPSer
         // a valid request
         var reshead = res.head
         reshead.headers.add(name: "date", value: RFC1123DateCache.shared.currentTimestamp())
+
         if let server = serverHeader {
             reshead.headers.add(name: "server", value: server)
         }
 
+        // add 'Connection' header if needed
+        let connectionHeaders = reshead.headers[canonicalForm: "connection"].map { $0.lowercased() }
```
weissi: you can make your life easier by just doing [`reqhead.isKeepAlive`](https://apple.github.io/swift-nio/docs/current/NIOHTTP1/Structs/HTTPRequestHead.html#/s:8NIOHTTP115HTTPRequestHeadV11isKeepAliveSbv) which will do a similar thing but also keep in mind if it's HTTP/1.0 (non-persistent which is like `connection: close` in HTTP/1.1) or HTTP/1.1 (defaults to persistent aka. keep-alive)

Sources/HTTP/Responder/HTTPServer.swift
```diff
@@ -236,10 +236,26 @@ private final class HTTPServerHandler<R>: ChannelInboundHandler where R: HTTPSer
...
+
+        if !connectionHeaders.contains("keep-alive") && !connectionHeaders.contains("close") {
```
weissi: I'd replace the whole body of this commit with

    reshead.headers.add(name: "Connection", value: reqhead.isKeepAlive ? "keep-alive" : "close")

NIO's [example](https://github.com/apple/swift-nio/blob/master/Sources/NIOHTTP1Server/main.swift#L43-L53) make this slightly more complicated to not even set the header if we match the default of the protocol but that's probably unnecessary.

enricenrich: But with the `connectionHeaders` we check if the response contains any `connection` header so we don’t duplicate it. Shouldn’t we keep it?

This said, agree on simplifying the rest of the code with just:

```
reshead.headers.add(name: "Connection", value: reqhead.isKeepAlive ? "keep-alive" : "close")
```

weissi: @enricenrich apologies, totally missed that. That dupe checking looks good then :).

joscdk:
@tanner0101 should we get this merged and tagged?

penny-coin:
Hey @enricenrich, you just merged a pull request, have a coin! 

You now have 1 coins.

PR #321, opened 2018-12-13
[HTTP-Pipelining] Test pipelined requests maintain order

t089:
This adds a simple test that asserts that pipelined requests are always responded to in the correct order if `supportPipelining: true` is set.


t089:
I've added a test that triggers the assertion in debug builds even if pipelining is enabled.

https://github.com/vapor/http/blob/939d1f93bcda7ad5295f1c2745d2eb962e63c2d7/Sources/HTTP/Responder/HTTPServer.swift#L150

PR #323, opened 2018-12-28
Fix missing port number in Host HTTP header

adtrevor:
This fixes an issue where the port wasn't included in requests Host header. Port is now included if it is not the default port for the scheme.


adtrevor:
The test that didn't pass was caused by amazon.com sending gzip compressed data while HTTPClient doesn't support it. I added an HTTP header specifying to the server that we don't support compression.

adtrevor:
Closing, too many changes since original PR. Will fix this in a new PR.

PR #359, opened 2019-04-02
Make the `HTTPMessageDecoder.decode` extension method public

MrMage:
Also adds support for decoding the JSON API media type (see https://jsonapi.org/format/). (Requires https://github.com/vapor/core/pull/202.)


MrMage:
It seems to already be public there, that's why I didn't file a second PR.

MrMage:
Renamed `jsonApi` to `jsonAPI` here as well.

penny-coin:
Hey @MrMage, you just merged a pull request, have a coin! 

You now have 21 coins.

PR #364, opened 2019-04-22
added .https scheme with TLSConfiguration parameter

popaaaandrei:
Nothing fancy, just made public a HTTPScheme generator that accepts custom TLSConfiguration. The reason is that we need to connect to a legacy API with client certificates.

Thanks


popaaaandrei:
customHTTPS already exists, there is no need for this PR.

PR #380, opened 2021-10-06
Add application/wasm HTTPMediaType

PatrickPijnappel:
Adds the `application/wasm` HTTP media type for `.wasm` files. Especially now with SwiftWasm gearing up this is relevant as [WebAssembly.instantiateStreaming(…)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WebAssembly/instantiateStreaming) requires the correct MIME type to be set.


0xTim (approved):
Thanks!

0xTim:
@PatrickPijnappel Did you mean to submit that here? HTTP is long deprecated and not used with Vapor 4

PatrickPijnappel:
@0xTim Ah I see, no didn't mean to 😅

Created one in the main repo instead: https://github.com/vapor/vapor/pull/2697

